PR #4, opened 2024-08-07
Prc 20 adding api endpoint

prabash-moj:
### **Description :** 
Add a get endpoint to get prisoner by prisoner id by internally making a api call to external prisoner search api to get the prisoner details,

### **Test evidence:**
image


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/PrisonerController.kt
```diff
@@ -0,0 +1,22 @@
...
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.client.prisonersearch.Prisoner
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.service.PrisonerService
+
+@Tag(name = "Get Prisoner")
+@RestController
+@RequestMapping(value = ["contacts"], produces = [MediaType.APPLICATION_JSON_VALUE])
+class PrisonerController(private val prisonerService: PrisonerService) {
+  @GetMapping(value = ["/prisoner/{prisonerNumber}"], produces = [MediaType.APPLICATION_JSON_VALUE])
+  @PreAuthorize("hasAnyRole('PRISONER_SEARCH')")
+  fun getPrisoner(
```
ReedSoftware, line 19: Should we have `@ResponseStatus`, `@Operation` and `@ApiResponses` as we do in the CSIP api?

rosariodiaferia-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/PrisonerController.kt
prabash-moj: Thanks James for the review, For this PR I would keep it simple as this is more likely to get dropped or updated when we play functional requirements , this is just a technical story aiming to get basic set up only.

PR #8, opened 2024-08-20
WIP : PRC-41 : API endpoint to provide a list of contacts associated with a…

prabash-moj:
… prisoner


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/entity/PrisonerContactViewEntity.kt
```diff
@@ -0,0 +1,43 @@
+package uk.gov.justice.digital.hmpps.hmppscontactsapi.entity
+
+import jakarta.persistence.Column
+import jakarta.persistence.Entity
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import java.time.LocalDate
+
+@Entity
+@Table(name = "prisoner_contact_view")
```
prabash-moj, line 10: will replace with the content from PR 7

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/repository/PrisonerContactViewRepository.kt
```diff
@@ -0,0 +1,13 @@
+package uk.gov.justice.digital.hmpps.hmppscontactsapi.repository
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.entity.PrisonerContactViewEntity
+
+@Repository
+interface PrisonerContactViewRepository : JpaRepository<PrisonerContactViewEntity, Long> {
+
```
prabash-moj, line 9: will replace with the content from PR 7

src/main/resources/migrations/common/V2024.08.02.1__baseline_tables.sql
```diff
@@ -286,6 +286,51 @@ CREATE TABLE reference_codes
...
+    c.first_name AS forename,
+    c.middle_name,
+    c.date_of_birth,
+    '' AS relationship_code, -- Placeholder as relationship details are not present in the current schema
+    '' AS relationship_description, -- Placeholder as relationship details are not present in the current schema
+    CASE
+        WHEN EXISTS (
+            SELECT 1 FROM contact_address ca
+            WHERE ca.contact_id = c.contact_id AND ca.primary_address = true
+        ) THEN ARRAY(
```
prabash-moj, line 304: will replace with the content from PR 7

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/mapping/ContactMappers.kt
```diff
@@ -0,0 +1,19 @@
...
+fun PrisonerContactSummary.toModel() = PrisonerContact(
+  prisonerContactId = prisonerContactId,
+  contactId = contactId,
+  surname = lastName,
+  forename = firstName,
+  middleName = middleName,
+  dateOfBirth = dateOfBirth!!,
+  relationshipCode = relationshipType,
+  relationshipDescription = relationshipDescription!!,
+  addresses = "TODO: map addresses",
```
prabash-moj, line 15: WIP: Looking into this now

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/PrisonerContactsController.kt
```diff
@@ -0,0 +1,78 @@
...
+  )
+  @GetMapping(value = ["/prisoner/{prisonerNumber}"], produces = [MediaType.APPLICATION_JSON_VALUE])
+  @PreAuthorize("hasAnyRole('ROLE_CONTACTS_ADMIN')")
+  fun getAllContacts(
+    @PathVariable prisonerNumber: String,
+    @RequestParam(name = "active", defaultValue = "true") active: Boolean,
+  ): ResponseEntity<List<PrisonerContact>> {
+    logger.info("Received request to fetch all contacts for prisoner number: $prisonerNumber with active status: $active")
+    val contacts = contactService.getAllContacts(prisonerNumber, active).toModel()
+    logger.info("Returning {} contacts", contacts.size)
```
timharrison-moj, line 75: Can use the shortened version of this function, like 

```
getAllContact( args..) = contactService.getAllContacts(args).toModel()
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/PrisonerContactsController.kt
```diff
@@ -0,0 +1,78 @@
...
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.service.ContactService
+
+@Tag(name = "Prisoner Contacts Controller")
+@RestController
+@RequestMapping("/prisoner-contacts", produces = [MediaType.APPLICATION_JSON_VALUE])
+class PrisonerContactsController @Autowired constructor(
+  private val contactService: ContactService,
+) {
+  private val logger = LoggerFactory.getLogger(PrisonerContactsController::class.java)
+
```
timharrison-moj, line 31: Logger should be companion object like ```companion object {
    private val log = LoggerFactory.getLogger(this::class.java)
  }```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.entity.PrisonerContactSummary
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.repository.PrisonerContactRepository
+
+@Service
+class ContactService @Autowired constructor(
+  private val prisonerContactRepository: PrisonerContactRepository,
```
timharrison-moj, line 11: No Autowire required, and logger in a companion object.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/PrisonerContactsController.kt
```diff
@@ -0,0 +1,78 @@
...
+    ],
+  )
+  @GetMapping(value = ["/prisoner/{prisonerNumber}"], produces = [MediaType.APPLICATION_JSON_VALUE])
+  @PreAuthorize("hasAnyRole('ROLE_CONTACTS_ADMIN')")
+  fun getAllContacts(
+    @PathVariable prisonerNumber: String,
+    @RequestParam(name = "active", defaultValue = "true") active: Boolean,
+  ): ResponseEntity<List<PrisonerContact>> {
+    logger.info("Received request to fetch all contacts for prisoner number: $prisonerNumber with active status: $active")
+    val contacts = contactService.getAllContacts(prisonerNumber, active).toModel()
```
timharrison-moj, line 74: Mapping between repo and external types should be done in the service layer please, not in the controller (resource) layer.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
```diff
@@ -0,0 +1,21 @@
...
+) {
+  private val logger = LoggerFactory.getLogger(ContactService::class.java)
+
+  fun getAllContacts(prisonerNumber: String, active: Boolean): List<PrisonerContactSummary> {
+    logger.info("Fetching all contacts")
+    val contacts = prisonerContactRepository.findPrisonerContacts(prisonerNumber, active)
+    logger.info("Found {} contacts", contacts.size)
+    return contacts
```
timharrison-moj, line 19: Can be shorthand version of function here..

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/PrisonerContact.kt
```diff
@@ -0,0 +1,37 @@
...
+
+  @Schema(description = "The relationship code between the prisoner and the contact", example = "FRI")
+  val relationshipCode: String,
+
+  @Schema(description = "The description of the relationship", example = "Friend")
+  val relationshipDescription: String,
+
+  @Schema(description = "Addresses associated with the contact, one of which should be primary")
+  val addresses: String,
+
```
timharrison-moj, line 34: The addresses probably shouldn't be here - there is one primary address, so the fields can be directly in the response object.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/PrisonerContactServiceTest.kt
```diff
@@ -0,0 +1,114 @@
...
+      awareOfCharges = false,
+      nextOfKin = false,
+      emergencyContact = false,
+      comments = "No comments",
+    )
+
+    val contacts = listOf(prisonerContactSummary1, prisonerContactSummary2)
+
+    Mockito.`when`(prisonerContactRepository.findPrisonerContacts("somePrisonerNumber", true)).thenReturn(contacts)
+
```
timharrison-moj, line 105: Can import Mockito.when directly - don't need the prefix.

timharrison-moj (commented):
Some comment added - mostly down to the "usual" way of doing things in MOJ teams - as you look at more of the services, you will pick up more of the conventions.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/PrisonerContact.kt
timharrison-moj: The exposed model for PrisonerContact doesn't need all the fields that the repository method returns, so where the mapper is used to produce the model object, perhaps here is where we can lose some of the detail, and return only the values required. Naming of objects here is going to get confusing, so I will think about good names ... PrisonerContactSummary is probably a betting name for the model object returned to the user, whilst PrisonerContactDetail might be better for what the view returns from the DB.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/PrisonerContactsController.kt
```diff
@@ -0,0 +1,78 @@
...
+        responseCode = "403",
+        description = "Forbidden, requires an appropriate role",
+        content = [
+          Content(
+            mediaType = "application/json",
+            schema = Schema(implementation = ErrorResponse::class),
+          ),
+        ],
+      ),
+    ],
```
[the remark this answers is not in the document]
prabash-moj: My understanding is UI will call prisoner-search directly to get the list of prisoners, then use a valid prisoner id to make this call, so we do not need to validate it again here by calling prisoner-search api, we will check only for basic null validation etc, is that different to what you think?

But agree there should be 404 for not found contact for a id and status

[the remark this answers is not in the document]
timharrison-moj: We shouldn't rely on UI checking for API features -  this endpoint may be called by clients other than our own UI. Best thing is to re-check that the prisoner is valid before attempting to get contacts. Should be a simple service-layer check with prisoner search. I can add afterwards though - if you focus on getting this first iteration in place.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/PrisonerContactsController.kt
```diff
@@ -0,0 +1,78 @@
...
+          Content(
+            mediaType = "application/json",
+            schema = Schema(implementation = ErrorResponse::class),
+          ),
+        ],
+      ),
+      ApiResponse(
+        responseCode = "404",
+        description = "The Contact was not found.",
+        content = [
```
timharrison-moj, line 63: This is for when the Prisoner is not found?

timharrison-moj (commented):
This is getting there now. Couple of minor comments left, and just need the address details adding into the response and we'll be at the first iteration.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/entity/PrisonerContactDetail.kt
```diff
@@ -9,7 +9,7 @@ import java.time.LocalDate
 @Entity
 @Immutable
 @Table(name = "v_prisoner_contacts")
-data class PrisonerContactSummary(
+data class PrisonerContactDetail(
   @Id
```
timharrison-moj, line 13: Thanks!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/mapping/ContactMappers.kt
```diff
@@ -0,0 +1,33 @@
...
+    area = this.area ?: "",
+    cityCode = this.cityCode ?: "",
+    countyCode = this.countyCode ?: "",
+    postCode = this.postCode ?: "",
+    countryCode = this.countryCode ?: "",
+    approvedVisitor = this.approvedVisitor,
+    nextOfKin = this.nextOfKin,
+    emergencyContact = this.emergencyContact,
+    awareOfCharges = this.awareOfCharges,
+    comments = this.comments ?: "",
```
timharrison-moj, line 29: V. good. That should do it.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/PrisonerContactSummaryServiceTest.kt
```diff
@@ -0,0 +1,145 @@
+package uk.gov.justice.digital.hmpps.hmppscontactsapi.service
+
+import jakarta.persistence.EntityNotFoundException
```
timharrison-moj, line 3: The name of this test file should really match the service name + "Test" - so its clear it contains tests for that service, though sometimes there are several test files for one service. Just a note for next time - again, its a convention type thing.

timharrison-moj: So really, should be ContactsServiceTest.kt

timharrison-moj (approved):
LGTM.

PR #15, opened 2024-08-22
PRC-52: Integrate get contact with create contact

james-baker-moj:
Create now returns the created contact as well a Location header. The tests now use get instead of the repository directly. Also, some minor swagger tidying for common responses.
Also also, I created a test API client so we can put common methods we're going to use across other integration tests, e.g., creating and getting a contact.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactController.kt
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.hmpps.hmppscontactsapi.resource
 
 import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.Parameter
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactController.kt
```diff
@@ -39,24 +42,14 @@ class ContactController(val contactService: ContactService) {
   )
   @ApiResponses(
     value = [
-      ApiResponse(responseCode = "201", description = "Created the contact successfully"),
       ApiResponse(
-        responseCode = "401",
-        description = "Unauthorised, requires a valid Oauth2 token",
+        responseCode = "201",
+        description = "Created the contact successfully",
+        headers = [ Header(name = "Location", description = "The URL where you can load the contact", example = "/contact/123456")],
```
prabash-moj, line 48: I think this formatting ok here, because no build failures due to formatting? break into multiple lines to match with others

prabash-moj (approved):

PR #18, opened 2024-09-04
PRC-58: Add is over eighteen to crete and get contact flow

james-baker-moj:
Add support for entering an estimated date of birth (DOB) if the DOB is not known. Return the isOverEighteen flag by either

* Calculating it if there is a DOB
* Using the value supplied when creating the contact 
* Default to DO_NOT_KNOW

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
@@ -6,13 +6,17 @@ import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.hmppscontactsapi.entity.ContactEntity
 import uk.gov.justice.digital.hmpps.hmppscontactsapi.entity.ContactEntity.Companion.newContact
 import uk.gov.justice.digital.hmpps.hmppscontactsapi.model.request.CreateContactRequest
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.model.request.IsOverEighteen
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
```diff
@@ -44,7 +49,37 @@ class ContactService(
     firstName = entity.firstName,
     middleName = entity.middleName,
     dateOfBirth = entity.dateOfBirth,
+    isOverEighteen = mapIsOverEighteen(entity),
     createdBy = entity.createdBy,
     createdTime = entity.createdTime,
   )
+
+  private fun mapIsOverEighteen(entity: ContactEntity): IsOverEighteen {
```
james-baker-moj, line 57: I took this approach as it felt bad to return null for isOverEighteen when we have a date of birth and could calculate it. Happy to strip it if you think it's too confusing.

prabash-moj (approved):

src/main/resources/migrations/common/V2024.09.04.1__reset_sequences.sql
```diff
@@ -0,0 +1,7 @@
+---------------------------------------------------------------------------------------
+-- Support the estimated date of birth question. Is the contact over 18? yes, no, don't know
+----------------------------------------------------------------------------------------
+
+    alter table contact add column is_over_eighteen boolean;
+
```
timharrison-moj, line 6: We should probably move this into the main data model migration? Probably misleading in the reset_sequences migration. I have some DB changes to make after the sync migrate chat, so can do this then.

PR #20, opened 2024-09-06
PRC-84 Database changes ahead of sync and extra base data to match in DEV

timharrison-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/mapping/PrisonerContactMappers.kt
```diff
@@ -31,3 +34,21 @@ fun PrisonerContactSummaryEntity.toModel(): PrisonerContactSummary {
...
+    if (!this.dateOfBirth.isAfter(LocalDate.now().minusYears(18))) {
+      IsOverEighteen.YES
+    } else {
+      IsOverEighteen.NO
+    }
+  } else {
+    when (this.isOverEighteen) {
+      true -> IsOverEighteen.YES
+      false -> IsOverEighteen.NO
+      else -> {
```
james-baker-moj, line 49: Minor but I don't think the braces are needed here

james-baker-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/mapping/PrisonerContactMappers.kt
timharrison-moj: Thanks, I'll sort in next PR if that's ok?  Going to get this one in, but involves dropping DB objects and losing any data in there. But it replaces with lots more...

PR #21, opened 2024-09-09
PRC-55: Add contact relationship to create contact

james-baker-moj:
Added an optional relationship to the create contact request. This populates the prisoner_contact table and the contact will then be returned from the /prisoner/A1234BC/contact endpoint.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
@@ -1,14 +1,18 @@
 package uk.gov.justice.digital.hmpps.hmppscontactsapi.service
 
+import jakarta.persistence.EntityNotFoundException
 import jakarta.transaction.Transactional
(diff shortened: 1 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
```diff
@@ -54,6 +65,31 @@ class ContactService(
...
+
+  private fun mapContact(request: CreateContactRequest) =
+    newContact(
+      title = request.title,
+      lastName = request.lastName,
+      firstName = request.firstName,
+      middleName = request.middleName,
+      dateOfBirth = request.dateOfBirth,
+      isOverEighteen = mapIsOverEighteen(request),
+      createdBy = request.createdBy,
```
timharrison-moj, line 90: I've been putting mapping like functions in ./mappers ... though should probably live in `./services/mapping` - usually just extension functions which operate on a single entity or list, to convert between internal / DB types and presentation of them.

timharrison-moj (approved):
LGTM

PR #24, opened 2024-09-12
PRC-108: Add references code API

james-baker-moj:
Simple API to get by group code.

Lift and shift from BVLS with a few extra tests

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/entity/ReferenceCodeEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/entity/ReferenceCodeEntity.kt
@@ -1,0 +1,51 @@
+package uk.gov.justice.digital.hmpps.hmppscontactsapi.entity
+
+import jakarta.persistence.Entity
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import org.hibernate.Hibernate
+import java.time.LocalDateTime
+
+@Entity
+@Table(name = "reference_codes")
+class ReferenceCodeEntity(
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val referenceCodeId: Long = 0,
+
+  val groupCode: String,
+
+  val code: String,
+
+  val description: String,
+
+  val createdBy: String,
+
+  val createdTime: LocalDateTime = LocalDateTime.now(),
+) {
+  var amendedBy: String? = null
+    private set
+
+  var amendedTime: LocalDateTime? = null
+    private set
+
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ReferenceCodeService.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.mapping.toModel
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.model.response.ReferenceCode
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.repository.ReferenceCodeRepository
+
+@Service
+class ReferenceCodeService(private val referenceCodeRepository: ReferenceCodeRepository) {
+  fun getReferenceDataByGroup(groupCode: String): List<ReferenceCode> =
+    referenceCodeRepository.findAllByGroupCodeEquals(groupCode).toModel()
```
prabash-moj, line 11: would be good to cache it? [@Cacheable(](https://github.com/ministryofjustice/prison-api/blob/24ff92a56eac4f02e3101aa6cb1e0ed3a9ca46be/src/main/java/uk/gov/justice/hmpps/prison/repository/ReferenceDataRepository.java#L42)

prabash-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/entity/ReferenceCodeEntity.kt
```diff
@@ -0,0 +1,51 @@
...
+) {
+  var amendedBy: String? = null
+    private set
+
+  var amendedTime: LocalDateTime? = null
+    private set
+
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (other == null || Hibernate.getClass(this) != Hibernate.getClass(other)) return false
```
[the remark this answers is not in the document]
james-baker-moj: No I just noticed that myself and was stripping it to make it a data class.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ReferenceCodeService.kt
james-baker-moj: yeah - need to think about caching, maybe do in the front end instead?

prabash-moj: good point, for this, may be do it on the frontend as it is just static data and not heavy computations involved here, But if we need to reduce db calls and improve performance, scalability, increase security etc then best to go with hybrid approach?

PR #27, opened 2024-09-18
PRC-91: Basic CRUD endpoints for contact address sync

timharrison-moj:
This PR adds the endpoints that will be required by the 2-way sync for CONTACT_ADDRESS. We'll need similar for each of the other tables so that the sync service can perform CRUD operations when data changes in NOMIS.

NOTE: The sync service will have more relaxed validation rules than the UI endpoints to cater for duff data from NOMIS - however, I'm not sure just how bad it will be yet, so there is very little validation on the requests to create/update as yet. We will need to add some checks e.g. not more than 1 primary address set to true, or checks about minimal viable content.  These will come out during sync testing.

Also - this is not yet checking that values from reference data lists are acceptable either - still to do. 

The main point of this work is to get all of the structure in place and basic operations working - to be refined when we know more.


src/main/resources/migrations/common/V2024.08.19.1__reset_sequences.sql
```diff
@@ -2,12 +2,12 @@
 -- After loading reference data with specific sequence values, the sequences
 -- themselves need to be reset to the next number for generated ID values.
 --
-alter sequence if exists contact_contact_id_seq restart with 29;
-alter sequence if exists contact_address_contact_address_id_seq restart with 13;
+alter sequence if exists contact_contact_id_seq restart with 16;
```
timharrison-moj, line 5: This caught me out in testing - the base data was loading 15 rows, but the sequence next val for tests was set to 13.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/entity/ContactAddressEntity.kt
```diff
@@ -1,29 +1,67 @@
...
+
+  var createdBy: String,
+
+  @CreationTimestamp
+  val createdTime: LocalDateTime = now(),
+) {
+  var amendedBy: String? = null
+  var amendedTime: LocalDateTime? = null
+
+  override fun equals(other: Any?): Boolean {
```
james-baker-moj, line 55: equals and hashcode not needed as it's a `data class`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/mapping/ContactAddressMappers.kt
```diff
@@ -0,0 +1,51 @@
+package uk.gov.justice.digital.hmpps.hmppscontactsapi.mapping
+
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.entity.ContactAddressEntity
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.model.ContactAddress
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.model.request.CreateContactAddressRequest
+
+fun ContactAddressEntity.toModel(): ContactAddress {
+  return ContactAddress(
+    contactAddressId = this.contactAddressId,
+    contactId = this.contactId!!,
```
james-baker-moj, line 10: unsafe?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/entity/ContactAddressEntity.kt
```diff
@@ -1,29 +1,67 @@
...
   @Id
-  val contactAddressId: Long,
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  var contactAddressId: Long,
 
-  val flat: String,
+  val contactId: Long? = null,
 
-  val property: String,
+  var addressType: String,
```
james-baker-moj, line 21: any reason to be var instead of val

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/SyncService.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    val contactAddress = contactAddressRepository.findById(contactAddressId)
+      .orElseThrow { EntityNotFoundException("Contact address with ID $contactAddressId not found") }
+
+    if (contact.contactId != contactAddress.contactId) {
+      logger.error("Contact address update specified for a contact not linked to this address")
+      throw ValidationException("Contact ID ${contact.contactId} is not linked to the address ${contactAddress.contactAddressId}")
+    }
+
+    with(contactAddress) {
```
james-baker-moj, line 72: If you used `.copy(primaryAddress = request.primaryAddress)` then you could keep the object immutable. I guess this is why they are `var`?

src/main/resources/migrations/common/V2024.08.19.1__reset_sequences.sql
james-baker-moj: not sure if it's possible but could you do a `select max(id)` here to avoid manual id stuff?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/SyncServiceTest.kt
```diff
@@ -0,0 +1,290 @@
...
+    street = "Main Street",
+    area = "Howarth",
+    cityCode = "LEEDS",
+    countyCode = "YORKS",
+    postCode = "LS13 4KD",
+    countryCode = "UK",
+    createdBy = "TEST",
+  )
+
+fun CreateContactAddressRequest.toEntity(contactAddressId: Long = 0) =
```
james-baker-moj, line 254: Could these be private or in a mappers file? It might be a bit confusing if they leak in to other tests.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/SyncService.kt
timharrison-moj: I had some issues with the .copy() this time around for updating the same entity, though have done this before, hence the mutable attribute vars. I will look around at previous, and other, projects to see how its handled there.
https://stackoverflow.com/questions/53510047/idiomatic-way-for-updating-repository-in-kotlin suggests its a viable way here - so I'll check what my issue was.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/entity/ContactAddressEntity.kt
timharrison-moj: Some (not all) attributes are mutable

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/SyncServiceTest.kt
timharrison-moj: Yes - could be private, so will update - or check not used in other places

timharrison-moj: Moved into the class as private functions.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/SyncService.kt
timharrison-moj: Just done it again, and no issues - minor changes only, and all attributes `val` again  - except the updatedBy and updatedTime, which do need to be modifiable.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/entity/ContactAddressEntity.kt
timharrison-moj: All val again now.

src/main/resources/migrations/common/V2024.08.19.1__reset_sequences.sql
timharrison-moj: Yes we can! Done.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/entity/ContactAddressEntity.kt
james-baker-moj: Never heard that one myself... is spring/JPA/hibernate using it somehow? 
One thing that concerns me is that I think I'd be surprised if `Entity(1, "FOO") === Entity(1, "BAR")`.
Hopefully tests would pick it up but then again if you had `assertThat(actual).isEqualTo(expected)` it would pass unexpectedly.

james-baker-moj: Also, our other entities don't do this and are working fine. Should probably be consistent at least.

james-baker-moj (approved):
Still not convinced on the equals/hashcode thing but otherwise LGTM.

PR #28, opened 2024-09-19
PRC-125: API for contact relationship

james-baker-moj:
API for linking contacts with a prisoner when the contact already exists.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactController.kt
@@ -13,6 +13,7 @@ import org.slf4j.LoggerFactory
 import org.springframework.data.domain.Pageable
 import org.springframework.data.domain.Sort.Direction
 import org.springframework.data.web.PageableDefault
+import org.springframework.http.HttpStatus
 import org.springframework.http.MediaType
 import org.springframework.http.ResponseEntity
 import org.springframework.security.access.prepost.PreAuthorize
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactController.kt
```diff
@@ -112,4 +114,46 @@ class ContactController(val contactService: ContactService) {
...
+  fun addContactRelationship(
+    @PathVariable("contactId") @Parameter(
+      name = "contactId",
+      description = "The id of the contact",
+      example = "123456",
+    ) contactId: Long,
+    @Valid @RequestBody relationshipRequest: AddContactRelationshipRequest,
+  ): ResponseEntity<Any> {
+    contactService.addContactRelationship(contactId, relationshipRequest)
+    return ResponseEntity.status(HttpStatus.CREATED).build()
```
timharrison-moj, line 157: Could be a Void return type from the service method, and annotate controller function with `@ResponseStatus(HttpStatus.CREATED)`?  Avoids the need to build a ResponseEntity yourself?

timharrison-moj (dismissed):
LGTM.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactController.kt
james-baker-moj: Good point, I was going to add a Location originally but didn't know what it would return so removed it. Will tidy up on next change if I remember....

james-baker-moj: Actually, I need to rebase so will sort now

timharrison-moj (approved):
LGTM

PR #32, opened 2024-09-25
PRC-41: Postgresql does not like the same syntax as H2 for resetting sequence next values

timharrison-moj:


james-baker-moj (approved):
LGTM

PR #40, opened 2024-10-01
PRC-147: Add country, county and city description

james-baker-moj:
Reference data for country changed to use NOMIS codes as per city and county.

Removed unused APIs to lookup country by various ISO codes and internal ids.

Changed baseline data to use the NOMIS codes.

Updated views to outer join on country, county and city reference data.

Return the descriptions from both search and prisoner contact list.


src/main/resources/migrations/common/V2024.08.02.2__baseline_data.sql
```diff
@@ -674,274 +674,267 @@ values
         (225, 'YID', 'Yiddish', 99, 'yid', 'yi', 'Yiddish'),
         (226, 'ZHA', 'Zhuang; Chuang', 99, 'zha', 'za', 'Zhuang; Chuang');
 
-
-Insert into country_reference (country_id, nomis_code, nomis_description,
-                                iso_alpha2, iso_alpha3, iso_numeric, iso_country_desc, display_sequence)
+insert into country_reference (nomis_code, nomis_description, display_sequence)
```
james-baker-moj, line 677: Re-imported from pre-prod

timharrison-moj (approved):
LGTM. Thanks for sorting.

PR #44, opened 2024-10-02
 PRC-46 : Allow entering hyphens and apostrophes for name

prabash-moj:
PRC-46 : Allow entering hyphens and apostrophes for name for contact search


james-baker-moj (approved):
LGTM

PR #48, opened 2024-10-04
PRC-173: Automatically clean and recreate the DB on changes applied

timharrison-moj:
This should stop us having to manually clean the DB when the migrations change.
Only applied in DEV for now.


james-baker-moj (approved):
LGTM

PR #52, opened 2024-10-07
PRC-150: Add email addresses to get contact response

james-baker-moj:


src/main/resources/migrations/common/R__v_contact_emails.sql
```diff
@@ -0,0 +1,21 @@
+--
+-- Creates a view over the contact_email and reference_codes tables to return a list of all phone numbers with descriptions
+-- for all codes.
```
timharrison-moj, line 3: Comment mentions phone numbers?  Can be updated in next PR though.

james-baker-moj: Yeah I had noticed that and fixed in next PR. forgot this wasn't merged yet. thanks

PR #59, opened 2024-10-15
PRC-202: Domestic status support

james-baker-moj:
Rename marital status to domestic status in line with designs.

Added reference data and ability to sort the reference data according to preference rather than alphabetical. Imported with display order from NOMIS and tweaked with Chris.

Return the domestic status and description from get contact by id.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/mapping/ReferenceCodeMappers.kt
```diff
@@ -7,7 +7,8 @@ fun ReferenceCodeEntity.toModel() = ReferenceCode(
   referenceCodeId = referenceCodeId,
   groupCode = groupCode,
   code = code,
+  displayOrder = displayOrder,
   description = description,
 )
 
-fun List<ReferenceCodeEntity>.toModel() = map { it.toModel() }.sortedBy { it.description }
```
james-baker-moj, line 13: This took me way too long to figure out why it wasn't returning in my sort order....

timharrison-moj: A little, hidden ordering lurking in a mapper...

PR #69, opened 2024-10-17
PRC-172: removed contact_website table and added UNKNOWN address type.

timharrison-moj:
Contact website - cannot e synced to NOMIS and no current usage for it.
Address types can be null in NOMIS data (for migrate) - so added an option of UNKNOWN - this will translate to null for NOMIS.


src/main/resources/migrations/common/V2024.08.02.1__baseline_tables.sql
```diff
@@ -76,7 +76,7 @@ CREATE TABLE contact_address
 (
     contact_address_id bigserial NOT NULL CONSTRAINT contact_address_id_pk PRIMARY KEY,
     contact_id bigint NOT NULL REFERENCES contact(contact_id),
-    address_type varchar(12) NOT NULL, -- Reference code - ADDRESS_TYPE e.g. HOME, WORK, TEMPORARY
+    address_type varchar(12) NOT NULL, -- Reference code - ADDRESS_TYPE e.g. HOME, WORK, TEMPORARY, UNKNOWN
```
james-baker-moj, line 79: TEMPORARY is wrong? I know it's only a comment but FYI

james-baker-moj (approved):
LGTM

src/main/resources/migrations/common/V2024.08.02.1__baseline_tables.sql
timharrison-moj: Ok - will make a note and update in next PR, thanks.

PR #70, opened 2024-10-17
PRC-206 : Patch a contact

prabash-moj:
PRC-206 : Patch a contact


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/request/patch/PatchContactRequest.kt
```diff
@@ -0,0 +1,88 @@
...
+    example = "This contact has special dietary requirements.",
+    nullable = true,
+  )
+  var comments: String? = null,
+
+  @Schema(description = "The id of the user who updated the contact", example = "JD000001", nullable = true)
+  val updatedBy: String? = null,
+
+  @Schema(description = "The timestamp of when the contact was changed", example = "2024-01-01T00:00:00Z", nullable = true)
+  val updatedTime: LocalDateTime? = null,
```
james-baker-moj, line 87: This should be set by the back end not in the request

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/mapping/patch/PatchContactEntityMappers.kt
```diff
@@ -0,0 +1,64 @@
...
+    languageCode = this.languageCode,
+    nationalityCode = this.nationalityCode,
+    interpreterRequired = this.interpreterRequired,
+    comments = this.comments,
+    amendedBy = this.amendedBy,
+    amendedTime = this.amendedTime,
+  )
+}
+
+fun ContactEntity.patchRequest(
```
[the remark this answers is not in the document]
prabash-moj: It would make easier to unit test by extracting into a own unit, I also see the information hiding point of view it may not feel good to separate, 

This would also make extensible as we can extract individual field into its own mapper with validation etc?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/request/patch/PatchContactRequest.kt
prabash-moj: good catch, I was thinking about this but slipped it out,

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/request/patch/PatchContactRequest.kt
```diff
@@ -0,0 +1,88 @@
...
+
+  @Schema(description = "Whether an interpreter is required", example = "false", nullable = true)
+  var interpreterRequired: Boolean? = null,
+
+  @Schema(
+    description = "Additional comments about the contact",
+    example = "This contact has special dietary requirements.",
+    nullable = true,
+  )
+  var comments: String? = null,
```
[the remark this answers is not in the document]
prabash-moj: will rebase and resolve

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/mapping/patch/PatchContactEntityMappers.kt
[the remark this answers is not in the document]
prabash-moj: Moved into own unit to so it only perform patch action - with improved unit tests

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/resource/PatchContactIntegrationTest.kt
```diff
@@ -0,0 +1,183 @@
...
+    val request = aPatchContactRequest()
+    val contactId = contact.id
+    val contactReturnedOnPatch = testAPIClient.patchAContact(request, "/contact/$contactId")
+
+    assertExpectedResponse(contactReturnedOnPatch, contact, request)
+    assertSuccessfullyPatched(testAPIClient.getContact(contactId), contact, request)
+  }
+
+  @Test
+  fun `should successfully patch the updated by field when its the only field in the request`() {
```
james-baker-moj, line 88: I don't think this test is doing what it says its for. The language code starts as null, is set to null and is checked is null so it's not testing that language code is not changed.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/resource/PatchContactIntegrationTest.kt
```diff
@@ -0,0 +1,183 @@
...
+    with(contactReturnedOnPatch) {
+      assertThat(languageCode).isEqualTo(null)
+      assertThat(amendedBy).isEqualTo("JD000001")
+    }
+    assertExpectedResponse(contactReturnedOnPatch, contact, request)
+    assertSuccessfullyPatched(testAPIClient.getContact(contactId), contact, request)
+  }
+
+  @Test
+  fun `should successfully patch the request`() {
```
james-baker-moj, line 105: This seems to be the same as `should successfully patch the contact with all fields`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
```diff
@@ -127,4 +148,38 @@ class ContactService(
     phoneNumbers: List<ContactPhoneNumberDetails>,
   ) = addressPhoneNumbers.filter { it.contactAddressId == contactAddressId }
     .mapNotNull { addressPhone -> phoneNumbers.find { it.contactPhoneId == addressPhone.contactPhoneId } }
+
+  fun ContactEntity.patchRequest(
+    request: PatchContactRequest,
+  ): ContactEntity {
+    val changedContact = this.copy(
+      title = this.title,
```
james-baker-moj, line 156: All these fields being set to themselves is quite confusing. Just set only language code instead as that's the only one that can actually be changed.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/request/patch/util/PatchableDeserializer.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.JsonDeserializer
+import com.fasterxml.jackson.databind.JsonNode
+
+class PatchableDeserializer : JsonDeserializer<Patchable<*>>() {
+  override fun deserialize(p: JsonParser, ctxt: DeserializationContext): Patchable<*> {
+    val node: JsonNode = p.codec.readTree(p)
+    return if (node.isNull || node.isMissingNode) {
```
james-baker-moj, line 11: This looks wrong. We want nulls to be set to present surely?

james-baker-moj (changes requested):

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/resource/PatchContactIntegrationTest.kt
```diff
@@ -0,0 +1,183 @@
...
+    val contactReturnedOnPatch = testAPIClient.patchAContact(request, "/contact/$contactId")
+
+    with(contactReturnedOnPatch) {
+      assertThat(languageCode).isEqualTo(null)
+      assertThat(amendedBy).isEqualTo("JD000001")
+    }
+    assertExpectedResponse(contactReturnedOnPatch, contact, request)
+    assertSuccessfullyPatched(testAPIClient.getContact(contactId), contact, request)
+  }
+
```
james-baker-moj, line 103: There should be a test where languageCode is set to a value and then set back to null

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/resource/PatchContactIntegrationTest.kt
prabash-moj: Confusion due to merging some tests - noticed an issue with json serialization from the integration tess, Have replaced with actual json object as a value, fixed those missing tests

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
james-baker-moj: I thought you were going to remove the ones in the `copy`?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/resource/PatchContactIntegrationTest.kt
```diff
@@ -0,0 +1,209 @@
...
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@TestPropertySource(properties = ["feature.event.contacts-api.prisoner-contact.amended=true"])
+@TestPropertySource(properties = ["feature.events.sns.enabled=false"])
+class PatchContactIntegrationTest : H2IntegrationTestBase() {
+
+  lateinit var contact: GetContactResponse
+
+  @BeforeAll
+  fun setUpContact() {
+    setupTestApiClient()
```
james-baker-moj, line 27: I don't think you need to call this directly as it's already in a `@BeforeEach`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
prabash-moj: Ops , have moved this to its own sections but looks I have forgot to remove it from here, Removed all this now

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/request/patch/util/PatchableSerializer.kt
```diff
@@ -0,0 +1,24 @@
...
+class PatchableSerializer : JsonSerializer<Patchable<*>>() {
+
+  override fun serialize(value: Patchable<*>?, gen: JsonGenerator, serializers: SerializerProvider) {
+    if (value != null) {
+      when (value) {
+        is Patchable.Present<*> -> {
+          val presentValue = value.get()
+          if (presentValue != null) {
+            gen.writeObject(presentValue)
+          }
```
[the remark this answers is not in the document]
prabash-moj: This is partially only used in integration tests only, This need further digging to remove the field when the field is undefined, 
when null and value is present it is working as expected , there are integration tests to cover those scenarios too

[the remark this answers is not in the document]
prabash-moj: Ok this is the tricky bit, this is only used in Integration tests, do not work as expected when serialise objects into json object when the  Patchable.undefined() case, other scenarios are fine, 

I will drop this now and use the import org.openapitools.jackson.nullable.JsonNullable instead of sealed classes, 
Its not worth the effort anymore unfortunately :(

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/resource/PatchContactIntegrationTest.kt
prabash-moj: I think this is requird, there is no before each

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactController.kt
```diff
@@ -195,4 +202,45 @@ class ContactController(val contactService: ContactService) {
     pageable: Pageable,
     @ModelAttribute @Valid @Parameter(description = "Contact search criteria", required = true) request: ContactSearchRequest,
   ) = contactService.searchContacts(pageable, request)
+
+  @PatchMapping("/{contactId}")
+  @Operation(
+    summary = "Update an contact",
+    description = "Update an contact",
```
james-baker-moj, line 209: minor grammar, should be "a" instead of "an"

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/resource/PatchContactIntegrationTest.kt
james-baker-moj: it's not required. remove it and it'll work same as all the others. H2IntegrationTestBase and PostgresIntegrationTestBase both extend the same class IntegrationTestBase. If you "find usages" you'll see this and the base class is the only place it's used (it could be private even maybe?).

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/resource/PatchContactIntegrationTest.kt
```diff
@@ -0,0 +1,209 @@
...
+    val contactId = contact.id
+    val contactReturnedOnPatch = testAPIClient.patchAContact(request, "/contact/$contactId")
+
+    assertExpectedResponse(contactReturnedOnPatch, contact, request)
+    assertSuccessfullyPatched(testAPIClient.getContact(contactId), contact, request)
+  }
+
+  @Test
+  fun `should successfully patch the updated by field when its the only field in the request`() {
+    val existingContact = testAPIClient.getContact(1)
```
james-baker-moj, line 89: I'm a bit concerned this might impact other tests as the migrations are only run once on spring boot startup as far as I know and aren't reset for every test?
All the other tests that do updates do so on newly created contacts AFAIK. 
I appreciate it's not as easy as you can't set the language code on create. I think it might be better to do multiple updates in the same test (set it to a value and then another with undefined and check the value is still set).

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/resource/PatchContactIntegrationTest.kt
```diff
@@ -0,0 +1,209 @@
...
+
+    with(contactReturnedOnPatch) {
+      assertThat(languageCode).isEqualTo("ENG")
+      assertThat(amendedBy).isEqualTo("JD000001")
+    }
+  }
+
+  @Test
+  fun `should successfully patch the language code with a null value `() {
+    val existingContact = testAPIClient.getContact(2)
```
james-baker-moj, line 116: Same with this test

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactPatchServiceTest.kt
```diff
@@ -0,0 +1,318 @@
...
+
+      whenever(contactRepository.findById(contactId)).thenReturn(Optional.empty())
+
+      org.junit.jupiter.api.assertThrows<EntityNotFoundException> {
+        service.patch(contactId, patchRequest)
+      }
+    }
+  }
+
+  fun mapToResponse(request: PatchContactRequest, existingContact: ContactEntity): PatchContactResponse {
```
james-baker-moj, line 224: do all these functions need to be public?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactPatchServiceTest.kt
```diff
@@ -0,0 +1,318 @@
...
+    fun `should throw EntityNotFoundException if contact does not exist`() {
+      val contactId = 1L
+      val patchRequest = PatchContactRequest(
+        languageCode = JsonNullable.of("ENG"),
+        updatedBy = "system",
+      )
+
+      whenever(contactRepository.findById(contactId)).thenReturn(Optional.empty())
+
+      org.junit.jupiter.api.assertThrows<EntityNotFoundException> {
```
james-baker-moj, line 218: import?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactPatchServiceTest.kt
```diff
@@ -0,0 +1,318 @@
...
+      it.languageCode = languageCode
+      it.nationalityCode = "GB"
+      it.interpreterRequired = false
+      it.amendedBy = "admin"
+      it.amendedTime = LocalDateTime.of(2024, 1, 22, 0, 0, 0)
+    }
+    return existingContact
+  }
+
+  fun mapToEntity(request: PatchContactRequest, existingContact: ContactEntity): ContactEntity {
```
james-baker-moj, line 282: In fact, as they are all only used once I'd just inline them for readability in the test.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactPatchServiceTest.kt
```diff
@@ -0,0 +1,318 @@
...
+      assertThat(updatedContact.suspended).isEqualTo(originalContact.suspended)
+      assertThat(updatedContact.staffFlag).isEqualTo(originalContact.staffFlag)
+      assertThat(updatedContact.coronerNumber).isEqualTo(originalContact.coronerNumber)
+      assertThat(updatedContact.gender).isEqualTo(originalContact.gender)
+      assertThat(updatedContact.nationalityCode).isEqualTo(originalContact.nationalityCode)
+      assertThat(updatedContact.interpreterRequired).isEqualTo(originalContact.interpreterRequired)
+      assertThat(updatedContact.domesticStatus).isEqualTo(originalContact.domesticStatus)
+      assertThat(updatedContact.amendedTime).isAfter(originalContact.amendedTime)
+      // patched fields
+      assertThat(updatedContact.languageCode).isEqualTo(null)
```
james-baker-moj, line 140: `isNull`

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/resource/PatchContactIntegrationTest.kt
prabash-moj: This is causing confusion because of BeforeAll and beforeEach order of the lifecycle sequence, I will try to remove this duplication, May be use sql insert statement using JDBC template or something to create the contact , That would also resolve the data leak or unexpected results due to updated data on the h2 db.

james-baker-moj (approved):
LGTM

PR #71, opened 2024-10-17
PRC-177: Support postgres based tests

james-baker-moj:
Extend PostgresIntegrationTestBase when you want to run with postgres and H2IntegrationTestBase when you don't.

Added dedicated test but also switched search to use postgres.


src/test/resources/application-h2.yml
```diff
@@ -0,0 +1,13 @@
+spring:
+
+  datasource:
+    url: 'jdbc:h2:mem:contacts-db;MODE=PostgreSQL'
+    username: contacts
+    password: dummy
+
+  flyway:
+    locations: classpath:/migrations/common,classpath:/migrations/h2
```
timharrison-moj, line 9: Can see flyway migrations for h2 - here.

src/test/resources/application-test.yml
```diff
@@ -3,18 +3,6 @@ spring:
   jpa:
     show-sql: true
 
-  datasource:
-    url: 'jdbc:h2:mem:contacts-db;MODE=PostgreSQL'
-    username: contacts
-    password: dummy
-
-  flyway:
-    locations: classpath:/migrations/common,classpath:/migrations/test
```
timharrison-moj, line 12: This used to apply flyway migrations for the "test" profile

timharrison-moj (commented):
LGTM - but a question about postgresql and flyway..

src/test/resources/application-test.yml
james-baker-moj: The only migration is V2024.08.02.3__reset_identities_for_baseline_data_on_h2.sql which is for h2 only.
I'll rename the folder to make it more obvious

james-baker-moj: Oh I did rename the folder this is the old code 😂 . For postgres, the migrations in `main` reset the sequences so it's not needed.

timharrison-moj (approved):
LGTM

PR #72, opened 2024-10-18
Demo json nullable

james-baker-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactController.kt
```diff
@@ -128,6 +130,51 @@ class ContactController(val contactService: ContactService) {
...
+  )
+  @PreAuthorize("hasAnyRole('ROLE_CONTACTS_ADMIN')")
+  fun patchContact(
+    @PathVariable("contactId") @Parameter(
+      name = "contactId",
+      description = "The id of the contact",
+      example = "123456",
+    ) contactId: Long,
+    @Valid @RequestBody request: PatchContactRequest,
+  ): ResponseEntity<Any> {
```
james-baker-moj, line 162: Just returns 3 different codes for the purposes of demoing it. This logic would be moved to the service layer and update the correct fields.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/request/PatchContactRequest.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.hmppscontactsapi.model.request
+
+import io.swagger.v3.oas.annotations.media.Schema
+import jakarta.validation.constraints.Size
+import org.openapitools.jackson.nullable.JsonNullable
+
+class PatchContactRequest {
```
james-baker-moj, line 7: The weak point is we can't use data class. It has to be initialised to undefined and any mandatory vars need to be lateinit which is not ideal.

timharrison-moj: There is a broadly similar approach taken here - https://techblog.dac.digital/sealed-classes-to-rescue-the-patch-endpoints-in-kotlin-d5aa7c3e02a6 - using sealed classes, which does still retain the data class for the request, but has another downside - requiring customer JSON mappers. All solutions will be similar - allowing the 3-states on request (present and null, present and not null, not present). Swings and roundabouts which method to use..

timharrison-moj: Your example here is clear enough, and does the job.

PR #81, opened 2024-10-23
PRC-207 :  patch interpreter required

prabash-moj:
API - Support update interpreter required in PATCH /contact/{id} endpoint


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactPatchServiceTest.kt
```diff
@@ -244,7 +265,7 @@ class ContactPatchServiceTest {
       domesticStatus = existingContact.domesticStatus,
       languageCode = request.languageCode.get(),
       nationalityCode = existingContact.nationalityCode,
-      interpreterRequired = existingContact.interpreterRequired,
+      interpreterRequired = request.interpreterRequired.get(),
```
james-baker-moj, line 268: There should be tests for each field separately rather than testing it's equal to the request. This is a very weak assertion.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/resource/PatchContactIntegrationTest.kt
```diff
@@ -93,6 +94,43 @@ class PatchContactIntegrationTest : H2IntegrationTestBase() {
     }
   }
 
+  @Test
+  fun `should successfully patch the contact with interpreter required`() {
```
james-baker-moj, line 98: this seems like 3 tests in one

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/helper/TestAPIClient.kt
```diff
@@ -118,6 +118,18 @@ class TestAPIClient(private val webTestClient: WebTestClient, private val jwtAut
     .expectBody(ErrorResponse::class.java)
     .returnResult().responseBody!!
 
+  fun getBadResponseErrorsWithPatch(request: Any, uri: URI) = webTestClient.patch()
```
james-baker-moj, line 121: this should take the proper type `PatchRequest` and the id instead of URI

prabash-moj: aiming for reusability, we already have lot of methods just changing the url or type of the request , we ideally could have one method for each of the four operations here  ? It wouldn't affect the readability in the test as we will build the input parameters to reflect the request and url.

james-baker-moj: ok this is for any patch request at all that returns an error? I see. I guess it's ok.  We do have a lot of dupe test client code when testing errors.  I'm not a huge fan of the `Any`. Maybe could be a type parameter instead?

james-baker-moj (approved):
LGTM thanks, tests look much better now

PR #96, opened 2024-10-28
PRC-204 : API - add isStaff to get contact by id endpoint

prabash-moj:
PRC-204 : API - add isStaff to get contact by id endpoint

1) add isStaff into getContactResponse
2) refactor other fields to be consistant


james-baker-moj (approved):
LGTM

PR #100, opened 2024-10-31
PRC-212: Fix deletion of phone number when it's linked to an address

james-baker-moj:
Also checking in the trivy that's been there for a week or more now...


timharrison-moj (approved):
LGTM

PR #101, opened 2024-10-31
PRC-214: Get and create contact identity

james-baker-moj:
First of the contact identity CRUD endpoints. GET and POST.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/entity/ContactIdentityDetailsEntity.kt
```diff
@@ -15,9 +15,6 @@ data class ContactIdentityDetailsEntity(
   val identityTypeDescription: String?,
   val identityValue: String?,
   val issuingAuthority: String?,
-  val verified: Boolean,
-  val verifiedBy: String?,
-  val verifiedTime: LocalDateTime?,
   val createdBy: String,
```
timharrison-moj, line 18: Yeah, I agree - only keep what is essential / used at the time.

timharrison-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactIdentityController.kt
```diff
@@ -0,0 +1,119 @@
...
+    ) contactId: Long,
+    @PathVariable("contactIdentityId") @Parameter(
+      name = "contactIdentityId",
+      description = "The id of the contact identity",
+      example = "987654",
+    ) contactIdentityId: Long,
+  ): ResponseEntity<Any> {
+    return contactIdentityFacade.get(contactId, contactIdentityId)
+      ?.let { ResponseEntity.ok(it) }
+      ?: throw EntityNotFoundException("Contact identity with id ($contactIdentityId) not found for contact ($contactId)")
```
prabash-moj, line 117: can we throw this exception in the service layer instead of the controller, 

I know it's bit inconsistent at the moment, some have done it in the repository layer, 
I think the best place should be in the service layer.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/facade/ContactIdentityFacade.kt
```diff
@@ -0,0 +1,25 @@
...
+  private val outboundEventsService: OutboundEventsService,
+) {
+
+  fun create(contactId: Long, request: CreateIdentityRequest): ContactIdentityDetails {
+    return contactIdentityService.create(contactId, request).also {
+      outboundEventsService.send(OutboundEvent.CONTACT_IDENTITY_CREATED, it.contactIdentityId)
+    }
+  }
+
+  fun get(contactId: Long, contactIdentityId: Long): ContactIdentityDetails? {
```
prabash-moj, line 22: Get service call is not needed to go though the facade , it is not transactional and do not send domain events. so contrller directly accessing this would be fine?

james-baker-moj: I know what you're saying, I kept it this way just to be consistent and less risky that someone would call `service.create` instead of `facade.create` from controller accidentally. They both have the same signature so avoiding having controller and facade both injected reduces the risk.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactIdentityController.kt
james-baker-moj: I think i'll move it into the facade, it's nice the service can return an optional result for other use cases but take your point maybe controller not the best. I actually hate throw exceptions for this stuff and would rather handle with a return type but that's a topic for another day :)

PR #103, opened 2024-11-01
PRC-214: Update contact identity

james-baker-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactIdentityController.kt
```diff
@@ -75,6 +77,52 @@ class ContactIdentityController(private val contactIdentityFacade: ContactIdenti
...
+    ) contactId: Long,
+    @PathVariable("contactIdentityId") @Parameter(
+      name = "contactIdentityId",
+      description = "The id of the contact identity",
+      example = "987654",
+    ) contactIdentityId: Long,
+    @Valid @RequestBody request: UpdateIdentityRequest,
+  ): ResponseEntity<Any> {
+    return ResponseEntity.ok(contactIdentityFacade.update(contactId, contactIdentityId, request))
+  }
```
timharrison-moj, line 124: Not sure we need to the ResponseEntity.ok(xxx)? Can just return the ContactIdentityDetails directly as ok is assumed in the absence of exceptions being caught and handled.

timharrison-moj (approved):
LGTM. Minor suggestion to avoid the ResponseIdentity, but not a blocker.

PR #107, opened 2024-11-02
PRC-248: Migration and minor data model updates

timharrison-moj:
PLEASE DO NOT MERGE YET - needs the UI types updating in line with these changes - breaking changes in a few endpoints - migration, prisoner contacts, search and contact summary.

It looks a scary amount of change, but it boils down to these basic changes :

* For CONTACT - addition of the `remitter` column (we'd removed it, but Syscon tell me it could still be used).

* For CONTACT_RESTRICTION and PRISONER_CONTACT_RESTRICTION - there is an `entereredByStaffUsername` field in the migrated objects, which can apparently be different from the `createdBy` field - so added separately in the data model.

* For PRISONER_CONTACT 
   - columns `awareOfCharges` and `canBeContacted` are not used in NOMIS, so I've removed from our data model, and downstream objects and tests.
   - I've added a new field called `currentTerm` which is a boolean to indicate whether each relationship applies to the latest term in prison (the latest booking). For users, we should only allow the search, view, create and management of relationships where the currentTerm is true,  and everything else is historical data which will come over in migration, or expired in response to a new booking event notification.

* I've updated some of the views and endpoints to return only the relationships where the `currentTerm = true` - but still need to address the search contact feature. Relationships can still be active true or false, or expired, even within the current term.

* The MigrationRequest - I've updated several fields in the migration request (and downstream) to reflect the correct nullability as advised by Syscon. For example, the `addressType` code need not be supplied on addresses and could be null. Also, on employments, the corporate name and ID may not be supplied. I've also agreed with Syscon that the lists of relationships, phones, addresses, identities, and restrictions supplied in the migration of one contact are non-nullable i.e if there are no addresses present for a contact we'll get an empty list (not a null or absent list).

I'll update the UI now to reflect these changes


run-local.sh
```diff
@@ -20,7 +20,7 @@ export API_BASE_URL_HMPPS_AUTH=https://sign-in-dev.hmpps.service.justice.gov.uk/
 export $(cat .env | xargs)  # If you want to set or update the current shell environment e.g. system client and secret.
 
 # Run the application with stdout and local profiles active
-SPRING_PROFILES_ACTIVE=stdout,local ./gradlew bootRun
+SPRING_PROFILES_ACTIVE=stdout,local,auto-db-clean ./gradlew bootRun
```
timharrison-moj, line 24: This will re-clean the database for local running against a postgresql container.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/mapping/ContactsServiceMappers.kt
```diff
@@ -104,14 +104,17 @@ private fun newPrisonerContact(
...
+    relationshipType = relationshipType,
+    nextOfKin = nextOfKin,
+    emergencyContact = emergencyContact,
+    comments = comments,
+    createdBy = createdBy,
+    createdTime = LocalDateTime.now(),
+    active = true,
+    approvedVisitor = false,
+    currentTerm = true,
   )
```
timharrison-moj, line 119: Just using field names - so it reports better any missing from the constructor.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/request/migrate/MigrateContactRequest.kt
```diff
@@ -8,16 +8,16 @@ import java.time.LocalDate
...
   var createUsername: String? = null,
 
-  @Schema(description = "The date and time the record was last amended", nullable = true)
+  @Schema(description = "The date and time the record was last amended", nullable = true, example = "2022-10-01T16:45:45")
   var modifyDateTime: LocalDateTime? = null,
 
-  @Schema(description = "The username who last modified the row", nullable = true)
+  @Schema(description = "The username who last modified the row", nullable = true, example = "X999X")
   var modifyUsername: String? = null,
 )
```
timharrison-moj, line 22: Just adding examples so Swagger docs read more clearly.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/PostgresIntegrationTestBase.kt
```diff
@@ -12,7 +12,7 @@ import uk.gov.justice.digital.hmpps.hmppscontactsapi.integration.wiremock.Prison
 
 @ExtendWith(HmppsAuthApiExtension::class, PrisonerSearchApiExtension::class)
 @SpringBootTest(webEnvironment = RANDOM_PORT)
-@ActiveProfiles("test", "local-postgres")
+@ActiveProfiles("test", "local-postgres", "auto-db-clean")
 abstract class PostgresIntegrationTestBase : IntegrationTestBase() {
```
timharrison-moj, line 16: This re-cleans the database for integration tests using a postgresql testcontainer.

prabash-moj (dismissed):

james-baker-moj (approved):
Looked good to Prabash 😂

PR #109, opened 2024-11-05
PRC-248: migration column lengths

timharrison-moj:
* Reference codes set to 12
* Flat (in address) changed 20 -> 30 chars
* Phone number 240 -> 40 charts
* Extension number optional (was mandatory)


james-baker-moj (approved):
LGTM

PR #110, opened 2024-11-05
PRC-251: Add identity reference data

james-baker-moj:
Imported from NOMIS including inactive types. Added validation that the type is active and return whether the type is active in ContactIdentityDetails so the UI can disable the edit button.


timharrison-moj (approved):
LGTM

PR #111, opened 2024-11-05
PRC-255: PATCH dateOfBirth and estimatedIsOverEighteen

james-baker-moj:
Support setting dateOfBirth and estimatedIsOverEighteen in the PATCH contact request.

Minor refactoring of ContactEntity to make fields part of the data class definition so we only need to copy and change relevant fields instead of additional random vars.


timharrison-moj (approved):
LGTM.

PR #114, opened 2024-11-07
PRC-256  : API - PATCH for contact relationship

prabash-moj:
PRC-256 : API - PATCH for contact relationship

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
@@ -10,7 +10,6 @@ import uk.gov.justice.digital.hmpps.hmppscontactsapi.entity.ContactAddressPhoneE
 import uk.gov.justice.digital.hmpps.hmppscontactsapi.entity.ContactEntity
 import uk.gov.justice.digital.hmpps.hmppscontactsapi.mapping.toEntity
 import uk.gov.justice.digital.hmpps.hmppscontactsapi.mapping.toModel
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
```diff
@@ -68,13 +67,6 @@ class ContactService(
...
     contactSearchRepository.searchContacts(request, pageable).toModel()
 
-  @Transactional
-  fun addContactRelationship(contactId: Long, request: AddContactRelationshipRequest) {
-    validateRelationship(request.relationship)
-    getContact(contactId) ?: throw EntityNotFoundException("Contact ($contactId) could not be found")
-    prisonerContactRepository.saveAndFlush(request.relationship.toEntity(contactId, request.createdBy))
-  }
-
   private fun validateRelationship(relationship: ContactRelationship) {
```
james-baker-moj, line 70: This bit is now duplicated as it's still used by create contact and now there are 2 places creating relationships.  I get what you are doing by separating it but also the pattern you set for `Contact` was to have separate `PatchContactService`  that just handles the patching of the entity. Maybe we should leave this alone and follow the pattern for contact?

prabash-moj: I see your point, was a bit confusing and did not realise that there are two ways to create a relationship, 
I see now this is using the add relationship request which is shared, 
```
data class AddContactRelationshipRequest(
  @Schema(description = "A description of the contacts relationship to a prisoner", exampleClasses = [ContactRelationship::class])
  val relationship: ContactRelationship,
```

My initial thought was to separate Create and update the relationship into its controller and service, No problem leave this out 👍

PR #115, opened 2024-11-08
PRC-286: Add CRUD endpoints for contact email

james-baker-moj:
Add all CRUD endpoints for email.

Removed the type and primary columns for email as these are not required or supported in NOMIS. This meant we could also remove the contact email view.


timharrison-moj (approved):
LGTM
Very thorough, and simplifies emails to match just what NOMIS tells us.

PR #116, opened 2024-11-08
PRC-283: Aligning reference data codes/values with NOMIS

timharrison-moj:
Aligning:
- Contact type
- Address type
- Gender
- (some domestic status values left in tests)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/request/sync/CreateContactRequest.kt
```diff
@@ -54,7 +54,7 @@ data class CreateContactRequest(
   @Schema(description = "The coroner's number, if applicable", example = "CRN12345", nullable = true)
   var coronerNumber: String? = null,
 
-  @Schema(description = "The gender of the contact", example = "Male", nullable = true)
+  @Schema(description = "The gender of the contact", allowableValues = ["M", "F", "NK", "NS"], example = "M", nullable = true)
```
james-baker-moj, line 57: Spring doesn't enforce these right?  Just thinking we don't want to have to keep it in sync with the DB

james-baker-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/request/sync/CreateContactRequest.kt
timharrison-moj: No - not enforcing as far as I know. However, as discussed - I'll add a note in swagger to describe that its a reference code, what its key is, and give some examples.

PR #118, opened 2024-11-11
PRC-288: Add gender and description to get contact by id

james-baker-moj:
Also renamed GetContactResponse to ContactDetails to match other "get by id" domain objects


timharrison-moj (approved):
LGTM

PR #119, opened 2024-11-11
PRC-189 & PRC-190: Send event on create contact and relationship

james-baker-moj:
Also used the facade to combine the 2 contact services and handle the relationship creation as part of contact or on it's own.


timharrison-moj (approved):
LGTM

PR #121, opened 2024-11-11
PRC-256 : API - PATCH for contact relationship

prabash-moj:
PRC-256 : API - PATCH for contact relationship


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/facade/ContactFacade.kt
```diff
@@ -55,4 +56,12 @@ class ContactFacade(
   fun searchContacts(pageable: Pageable, request: ContactSearchRequest): Page<ContactSearchResultItem> {
     return contactService.searchContacts(pageable, request)
   }
+
+  fun patchRelationship(id: Long, prisonerContactId: Long, request: UpdateRelationshipRequest) {
+    return contactService.updateContactRelationship(id, prisonerContactId, request)
```
timharrison-moj, line 61: What does `id` relate to here?  The `prisonerContactId` uniquely identifies one relationship..

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
```diff
@@ -134,4 +138,95 @@ class ContactService(
     phoneNumbers: List<ContactPhoneDetails>,
   ) = addressPhoneNumbers.filter { it.contactAddressId == contactAddressId }
     .mapNotNull { addressPhone -> phoneNumbers.find { it.contactPhoneId == addressPhone.contactPhoneId } }
+
+  @Transactional
+  fun updateContactRelationship(contactId: Long, prisonerContactId: Long, request: UpdateRelationshipRequest) {
+    val contact = getPrisonerContactEntity(prisonerContactId)
+
```
timharrison-moj, line 145: I think the naming of variables is confusing here ^ .   It retrieves a `PrisonerContactEntity` (this is the relationship) into a variable called `contact` - which in our terms is the person / visitor. It gets very confusing if inconsistent terms are used.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/resource/UpdateContactRelationshipIntegrationTest.kt
```diff
@@ -0,0 +1,282 @@
...
+
+  private fun getRandomPrisonerCode(): String {
+    val letters = ('A'..'Z')
+    val numbers = ('0'..'9')
+
+    val firstLetter = letters.random()
+    val lastLetter = letters.random()
+    val numberPart = (1..4).map { numbers.random() }.joinToString("")
+
+    return "$firstLetter$numberPart$lastLetter"
```
timharrison-moj, line 244: Prisoner numbers are format `A9999AA` (two letters at the end)

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactServiceTest.kt
```diff
@@ -655,6 +659,402 @@ class ContactServiceTest {
...
+    fun before() {
+      prisonerContact = createPrisonerContact(contactId)
+    }
+
+    @Nested
+    inner class RelationshipType {
+
+      @Test
+      fun `should update the contact relationship type`() {
+        val relationShipTypeCode = "FRI"
```
timharrison-moj, line 679: Some the tests are repetitive - perhaps consider introducing a CsvSource or MethodSource to repeat the same test with different inputs?

timharrison-moj (commented):
Couple of questions/comments but getting there.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactServiceTest.kt
prabash-moj: I had the same thoughts and tried my best to reduce code duplication in this unit test, However it has been difficult further reduction due to it is now down to setting different field name of the optional object , CsvSource would be useful to set value to field,  but not setting different optional field, 

This is similar situation which we have in other patch unit /int tests,

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/facade/ContactFacade.kt
timharrison-moj: `Id` here is vague? Clearer as `contactId`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactService.kt
timharrison-moj: Better now, thanks.  Always worth putting a response in the PR though.. to describe what you've done to address the comment?

timharrison-moj (approved):
LGTM

PR #122, opened 2024-11-11
PRC-214 : match NOMIS id types order

prabash-moj:
PRC-214 : match NOMIS id types order

Test evidence :

Expected as per PREPROD:
image

Actual:

image


timharrison-moj (approved):
LGTM

PR #123, opened 2024-11-11
PRC-280: Dropping unused columns from CONTACT and performance-related changes

timharrison-moj:
This PR is dropping unused columns from the CONTACT table, which equates to the PERSONS table in NOMIS.

Columns dropped:

* coroner_no - never populated in NOMIS
* suspended - always N (false) in NOMIS
* active - always Y (true) in NOMIS
* placeOfBirth - always null in NOMIS.
* nationalityCode - always null in NOMIS.

Also - testing with volume revealed some performance tuning was needed - indexes, views.


james-baker-moj (approved):
LGTM

PR #124, opened 2024-11-12
PRC-159: Display none primary addresses

james-baker-moj:
For contact list and contact search we can show not just the primary address but apply the following rules to select an address, in order, provided it does not have an end date.

1. Primary address
2. Mail address
3. Latest start date
4. Latest creation date

Also add primary and mail flags to returned types to allow labelling the type of address


src/main/resources/migrations/common/R__v_contacts_with_primary_address.sql
```diff
@@ -40,7 +40,13 @@ select
     ca.no_fixed_address,
     ca.comments
   from contact c
-  left join contact_address ca ON ca.contact_id = c.contact_id AND ca.primary_address = true
+  left join contact_address ca ON ca.contact_address_id = (
+      select contact_address_id from contact_address ca1
+      where ca1.contact_id = c.contact_id
+      and end_date is null
+      order by ca1.primary_address desc, ca1.mail_flag desc, ca1.start_date desc nulls last, ca1.created_time desc
+      limit 1
```
timharrison-moj, line 48: Yay. Looks good to me.

timharrison-moj (approved):
LGTM

PR #127, opened 2024-11-13
PRC-287: Allow migrate and sync to provide the PERSON_ID as the primary key for contact_id

timharrison-moj:
This PR:
* Allows migrate and sync to provide a primary key
* Aligns PERSON_ID in NOMIS with CONTACT_ID in our service.
* Will generate CONTACT_ID in a new range (from 20m) for locally created contacts.
* Will check/prevent duplicates from being migrated and synced
* Renamed one of the duplicated `CreateContactRequest` objects (to SyncCreateContactRequest)
* Unit and Integration tests added and adjusted for the above changes 

NOTE: There is some JPA magic involved in allowing a @GenerateValue primary key to be overridden with a provided value - see the custom annotation @SequenceOrUseId (courtesy of Syscon - Andy had to do almost exactly the same change their side to allow the PERSON_ID to override the normal sequence for NOMIS).


james-baker-moj (approved):
LGTM

PR #132, opened 2024-11-15
PRC-194: Limit contact list to one email

james-baker-moj:
This caused spring to blow up if you had more than one email.


timharrison-moj (approved):
LGTM

PR #133, opened 2024-11-18
PRC-312: Accept duplicates in migration and add more detail to outbound events.

timharrison-moj:
This PR:
* Allows duplicate PERSON_ID in migrate - if exists, it deletes the old and recreates.
* Avoids a 409 conflict in migrate (but still does not allow duplicate person on sync)
* Adds a source system (DPS or NOMIS) to outbound events - set to DPS for all at present (next PR will change this)
* Adds a personReference attribute to all outbound events


james-baker-moj (approved):
LGTM

PR #138, opened 2024-11-19
PRC-306: Return prisoner contact id on add/create contact

james-baker-moj:
This is required to be able to link into journeys after creation was successful.


timharrison-moj (approved):
LGTM

PR #140, opened 2024-11-19
PRC-310:  Sync changes also produce domain events with source = NOMIS.

timharrison-moj:
This PR:
* Introduces a sync facade to sit between sync controllers and sync services
* The sync facade calls the sync services, and then produces domain events.
* Added additional attributes to domain events - source (DPS or NOMIS) and a person reference 
* Sync events also produce domain events - contact create, amend and delete initially - with source - NOMIS.
* Stub events - for integration tests - updated to check the additional event attributes.
* Affected tests updated - there were a lot of these!
* New tests added for Sync facade and domain events (source NOMIS) being emitted.


james-baker-moj (approved):
LGTM - thanks for updating all the integration tests as well. Will stand us in good stead going forward.

PR #142, opened 2024-11-20
PRC-310: Fix (attempted) for missing source on events

timharrison-moj:
Its to do with data classes inheriting from a parent class, and the parent class attributes not being visible to the ObjectMapper.


james-baker-moj (approved):
LGTM

PR #148, opened 2024-11-21
PRC-310: Adding sync-triggered domain events (source NOMIS) for phones, emails and identities

timharrison-moj:
Getting the events and facade/service tests first.
I then have some integration tests to extend and check domain events using stub events (to follow)


james-baker-moj (approved):
LGTM

PR #152, opened 2024-11-23
PRC-310: More sync events and corrections for handling prisoner contact restrictions

timharrison-moj:
This PR contains sync-initiated events and tests for:
* Contact addresses
* Contact restrictions
* Prisoner contacts (relationships)
* Prisoner contact restrictions (restrictions on a relationship)
* I found quite a few problems with the code for prisoner contact restrictions - so I will have a sweep through the other side of this, for handling UI requests relating to these.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/sync/SyncContactAddressService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/sync/SyncContactAddressService.kt
@@ -46,12 +46,6 @@ class SyncContactAddressService(
     return contactAddress.toModel()
   }
 
-  fun deleteContactAddress(contactAddressId: Long) {
-    contactAddressRepository.findById(contactAddressId)
-      .orElseThrow { EntityNotFoundException("Contact address with ID $contactAddressId not found") }
-    contactAddressRepository.deleteById(contactAddressId)
-  }
-
   fun createContactAddress(request: SyncCreateContactAddressRequest): SyncContactAddress {
     contactRepository.findById(request.contactId)
       .orElseThrow { EntityNotFoundException("Contact with ID ${request.contactId} not found") }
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/sync/SyncContactAddressService.kt
```diff
@@ -98,4 +92,12 @@ class SyncContactAddressService(
 
     return contactAddressRepository.saveAndFlush(changedContactAddress).toModel()
   }
+
+  // TODO: Should this delete the contact_phone rows that are specifically associated with this address?
+  fun deleteContactAddress(contactAddressId: Long): SyncContactAddress {
```
james-baker-moj, line 97: I assume so, yes. At the very least this will cause a constraint violation exception on contact address phone so you will need to delete that. What happens in NOMIS when you delete an address? I assume the phone numbers go too.

james-baker-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/sync/SyncContactAddressService.kt
timharrison-moj: Thanks - I'll ask Andy what happens in NOMIS.

timharrison-moj: Ok - Andy says the general principal is, if we receive a sync call to delete a parent object, we take out all child entities linked with it. And if we get a sync delete for an entity which does not exist, or its parent does not exist, we should return a 200 ok, not a 404. I'll create a ticket for this - as I think we would currently 404 it.

PR #160, opened 2024-11-27
PRC-334 : Tech - switch language to use reference_codes

prabash-moj:
PRC-334 : Tech - switch language to use reference_codes instead of custom language tables


src/main/resources/migrations/common/V2024.08.19.1__reset_sequences.sql
```diff
@@ -12,7 +12,7 @@ alter sequence if exists contact_address_contact_address_id_seq restart with 18;
 alter sequence if exists contact_email_contact_email_id_seq restart with 5;
 alter sequence if exists contact_identity_contact_identity_id_seq restart with 5;
 alter sequence if exists contact_phone_contact_phone_id_seq restart with 13;
-alter sequence if exists reference_codes_reference_code_id_seq restart with 82;
+alter sequence if exists reference_codes_reference_code_id_seq restart with 269;
```
james-baker-moj, line 15: I don't think we actually need this line anymore as the inserts use the sequence now

james-baker-moj (approved):
LGTM

PR #161, opened 2024-11-27
PRC-339: Create and update APIs for prisoner contact restrictions

james-baker-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/entity/PrisonerContactRestrictionEntity.kt
```diff
@@ -32,9 +33,9 @@ data class PrisonerContactRestrictionEntity(
...
-  val createdTime: LocalDateTime,
-) {
-  var amendedBy: String? = null
+  val createdTime: LocalDateTime = now(),
 
-  var amendedTime: LocalDateTime? = null
-}
+  val amendedBy: String? = null,
+
+  val amendedTime: LocalDateTime? = null,
```
prabash-moj, line 40: not related to this PR,  just for consistency I also noticed that there are few entities where this is not nullable.   `val amendedTime: LocalDateTime?,` may be worth getting them consistent across the project

prabash-moj (approved):

PR #162, opened 2024-11-27
PRC-336 : Tech - Limit the UI display to SOCIAL contacts only

prabash-moj:
PRC-336 : Tech - Limit the UI display to SOCIAL contacts only


james-baker-moj (approved):
LGTM - I think this is the only bit we need to do for now.

PR #172, opened 2024-12-03
PRC-74: Entered by for prisoner-contact restrictions

james-baker-moj:
Use the manage users API to lookup the entered by users name for prisoner-contact restrictions. Entered by is classed as the creating user if never updated and the last updated user otherwise.


timharrison-moj (commented):
LGTM

`Code looks code, but I don't see any URL env vars added to provide the manage-users-api URL to the service in helm_deploy, or mocked APIs for testing or as dependencies in health checks? Or are these present already?  `- ignore this - they were added in the last PR!

timharrison-moj (approved):
LGTM

PR #173, opened 2024-12-05
PRC-343: Update spring starter

james-baker-moj:
There were several related issues including string random utils changes, code generator (for code not even used) and the id generator for contacts which allows overriding with a specific ID when creating from NOMIS.

Removed the generated code.

Updated uses of RandomStringUtils to use the secure() methods.

Use identity column for Contact with default value. Split the entities into types with or without known id.


build.gradle.kts
```diff
@@ -68,54 +64,6 @@ kotlin {
 
 tasks {
   withType<KotlinCompile> {
-    dependsOn("buildPrisonerSearchApiModel")
     compilerOptions.jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_21
   }
-  withType<KtLintCheckTask> {
-    // Under gradle 8 we must declare the dependency here, even if we're not going to be linting the model
-    mustRunAfter("buildPrisonerSearchApiModel")
-  }
```
timharrison-moj, line 77: I think we use the prisoner search API model to get prisoner details? So need to be generating the types from its OpenAPI spec?

timharrison-moj: Not at present - but may be one to introduce if we start to use more data and need to understand the types better. Resolving this for now.

timharrison-moj (approved):
LGTM.

james-baker-moj:
Please hold merging until Grace has finished her testing

PR #174, opened 2024-12-06
PRC-323: Only allow primary and mail on one address

james-baker-moj:
Changed the create and update APIs for contact addresses to remove the primary and mail flags from other addresses if setting them on the subject. Also generates updated events for any other addresses updated.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/repository/ContactAddressRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/repository/ContactAddressRepository.kt
@@ -12,4 +12,30 @@ interface ContactAddressRepository : JpaRepository<ContactAddressEntity, Long> {
   @Modifying
   @Query("delete from ContactAddressEntity ca where ca.contactId = :contactId")
   fun deleteAllByContactId(contactId: Long): Int
+
+  @Modifying
+  @Query(
+    value = """
+      UPDATE contact_address 
+      SET primary_address = false 
+      WHERE primary_address = true
+      AND contact_id = :contactId 
+      RETURNING contact_address_id
+      """,
+    nativeQuery = true,
+  )
+  fun resetPrimaryAddressFlagForContact(contactId: Long): List<Long>
+
+  @Modifying
+  @Query(
+    value = """
+      UPDATE contact_address 
+      SET mail_flag = false 
+      WHERE mail_flag = true
+      AND contact_id = :contactId 
+      RETURNING contact_address_id
+      """,
+    nativeQuery = true,
+  )
+  fun resetMailAddressFlagForContact(contactId: Long): List<Long>
 }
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/repository/ContactAddressRepository.kt
```diff
@@ -12,4 +12,30 @@ interface ContactAddressRepository : JpaRepository<ContactAddressEntity, Long> {
...
   fun deleteAllByContactId(contactId: Long): Int
+
+  @Modifying
+  @Query(
+    value = """
+      UPDATE contact_address 
+      SET primary_address = false 
+      WHERE primary_address = true
+      AND contact_id = :contactId 
+      RETURNING contact_address_id
```
timharrison-moj, line 23: Should this also limit to where `currentTerm=true` ?  Applying to this term in prison, as their primary and mail flags might have been different 5, 10 or 15 years ago?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/repository/ContactAddressRepository.kt
```diff
@@ -12,4 +12,30 @@ interface ContactAddressRepository : JpaRepository<ContactAddressEntity, Long> {
...
+  @Query(
+    value = """
+      UPDATE contact_address 
+      SET mail_flag = false 
+      WHERE mail_flag = true
+      AND contact_id = :contactId 
+      RETURNING contact_address_id
+      """,
+    nativeQuery = true,
+  )
```
timharrison-moj, line 39: Postgresql specific?

timharrison-moj (approved):
LGTM - though a question on currentTerm?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/repository/ContactAddressRepository.kt
james-baker-moj: yes unfortunately H2 doesn't support this in the postgres specific dialect. I tried looking for an implementation agnostic way to do it but the only way is to load all the entities and update them but it's a lot messier. Happy to reconsider if you're not keen

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/repository/ContactAddressRepository.kt
james-baker-moj: I don't think we can link it to the term as it's against contact rather than prisoner-contact?

timharrison-moj: Yes, you're right. Not against relationship, so its fine.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/repository/ContactAddressRepository.kt
timharrison-moj: Ok. That's fine.

PR #180, opened 2024-12-11
PRC-323: Address type order

james-baker-moj:
Should be home, business and then work but default order is alphabetical on code so set the display sequence correctly.


timharrison-moj (approved):
LGTM

PR #185, opened 2024-12-16
PRC-349: Re-organise swagger docs

james-baker-moj:
Sort into Contacts, Prisoner Contact Relationships, Restrictions and Reference Data in that order with Sync and Migrate endpoints last.

Also added the server URLs to allow switching for testing and a good way of publicly documenting the URLs.


timharrison-moj (approved):
LGTM

PR #187, opened 2024-12-17
PRC-345 : Tech ticket - validate city, county & country codes in create/update address

prabash-moj:
PRC-345 : Tech ticket - validate city, county & country codes in create/update address


james-baker-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactAddressService.kt
```diff
@@ -114,6 +127,23 @@ class ContactAddressService(
...
+
+  private fun validateCountyCode(countyCode: String?) {
+    countyCode?.let { validateReferenceDataExists(it, COUNTY) }
+  }
+
+  private fun validateCityCode(cityCode: String?) {
+    cityCode?.let { validateReferenceDataExists(it, CITY) }
+  }
+
+  private fun validateReferenceDataExists(code: String, groupCode: String) =
```
james-baker-moj, line 142: it's on my list to make group code an enum and maybe move the actual validation to the reference service as we've ended up with a lot of these

PR #188, opened 2024-12-18
PRC-140: Linked prisoners API

james-baker-moj:
API to get prisoners linked to a contact, this can return social and official contacts but the other uses of v_prisoner_contacts are still limited to Social only.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/LinkedPrisonersService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/LinkedPrisonersService.kt
@@ -1,0 +1,38 @@
+package uk.gov.justice.digital.hmpps.hmppscontactsapi.service
+
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.model.response.LinkedPrisonerDetails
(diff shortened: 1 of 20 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/LinkedPrisonersService.kt
```diff
@@ -0,0 +1,38 @@
...
+              middleNames = prisoner.middleNames,
+              relationships = summaries.map { summary ->
+                LinkedPrisonerRelationshipDetails(
+                  prisonerContactId = summary.prisonerContactId,
+                  contactType = summary.contactType,
+                  contactTypeDescription = summary.contactTypeDescription,
+                  relationshipCode = summary.relationshipType,
+                  relationshipDescription = summary.relationshipDescription,
+                )
+              },
```
timharrison-moj, line 33: This looks like it will call prisoner search once for each linked prisoner? It's probably ok (very quick to call) but longer term, I wonder about using an alternative endpoint on prisoner search which accepts a list of prisoner numbers and gets all their details at once?

timharrison-moj (approved):
LGTM. Minor comment, for later consideration if we end up calling prisoner search 20-30 times for some people with lots of links.

PR #189, opened 2024-12-18
PRC-140: Fix migrations

james-baker-moj:
Drop views before creating them rather than using CREATE OR REPLACE. This allows re-name and re-order of columns which does not work with CREATE OR REPLACE. These migrations are only re-run if the checksum changes.


timharrison-moj (approved):
LGTM

PR #190, opened 2024-12-18
PRC-352-Tech ticket - SyncUpdateContactIdentityRequest.issuingAuthority is mandatory but should be optional

prabash-moj:
PRC-352-Tech ticket - SyncUpdateContactIdentityRequest.issuingAuthority is mandatory but should be optional


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/request/sync/SyncUpdateContactIdentityRequest.kt
```diff
@@ -15,7 +15,7 @@ data class SyncUpdateContactIdentityRequest(
   val identityValue: String,
 
   @Schema(description = "Issuing authority", example = "DVLA")
-  val issuingAuthority: String,
+  val issuingAuthority: String?,
```
timharrison-moj, line 19: `nullable = true` on @Schema attributes?

timharrison-moj: Any `maxLength` attribute?

timharrison-moj (commented):
LGTM. Minor comment on adding nullable. 
I think the max length would be nice, but not vital for now as these values are coming from NOMIS and we're aligned on column sizes here.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/request/sync/SyncUpdateContactIdentityRequest.kt
prabash-moj: added both max length and nullable for swagger.

timharrison-moj (approved):
LGTM

PR #192, opened 2024-12-20
PRC-250: PNC validation

james-baker-moj:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/util/PNCValidatorTest.kt
```diff
@@ -0,0 +1,705 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.CsvSource
+
+class PNCValidatorTest {
+
+  @ParameterizedTest
+  @CsvSource(
+    value = [
+      "2008/0056560Z",
```
james-baker-moj, line 12: I copied all the test cases from CPR. Not sure they are all really required but I also don't know why they had them. Could probably be cut down significantly. One for the back burner when I get annoyed how long the test takes to run.

PR #193, opened 2024-12-20
PRC-365: Add additional event id into contact address phones

timharrison-moj:
For contact address phones, and in particular the delete operations, Syscon are unable to retrieve the row by ID after it has been deleted, so need a second ID (the contactAddressId) to work with.


andymarke (approved):
LGTM

PR #198, opened 2024-12-23
PRC-366: Ordering of deletes for sync contact address phone

timharrison-moj:
Reorder deletes to match the way this is done in the non-sync service.


andymarke (approved):
LGTM

PR #202, opened 2025-01-03
PRC-348 : Tech ticket - Add PATCH API for address

prabash-moj:
The manage address journeys should be individual “bounce” pages so it would be better to have a PATCH API that allows us to update only the relevant fields to each page (type, address lines, meta data).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactAddressController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactAddressController.kt
@@ -14,6 +14,7 @@ import org.springframework.http.ResponseEntity
 import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.web.bind.annotation.DeleteMapping
 import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PatchMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.PutMapping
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactAddressController.kt
```diff
@@ -111,6 +113,44 @@ class ContactAddressController(private val contactAddressFacade: ContactAddressF
     request: UpdateContactAddressRequest,
   ) = contactAddressFacade.update(contactId, contactAddressId, request)
 
+  @PatchMapping("/{contactAddressId}", consumes = [MediaType.APPLICATION_JSON_VALUE])
```
timharrison-moj, line 116: When this goes out to DEV, will anything in UI still call the POST endpoint? i.e. will the update endpoint be redundant?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactAddressService.kt
```diff
@@ -108,6 +110,69 @@ class ContactAddressService(
...
+    }
+
+    request.mailFlag.ifPresent {
+      if (it && !existing.mailFlag) {
+        updatedAddressIds += contactAddressRepository.resetMailAddressFlagForContact(contactId)
+      }
+    }
+
+    if (contact.contactId != existing.contactId) {
+      logger.error("Contact address update specified for a contact not linked to this address")
```
timharrison-moj, line 134: This probably wants to be placed before doing any of the updates for primary/mail addresses, and probably before the city/county/country code validation?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/ContactAddressController.kt
prabash-moj: I had the same thoughts,  we can add a tech ticket to move the UI to new patch endpoint , better to clean up than keeping the duplicate code as it pretty much does the same thing but this has added features,

timharrison-moj: ok - its fine to leave both in the API until the UI has been switched over. Thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/service/ContactAddressService.kt
prabash-moj: I have updated this now

timharrison-moj: Thanks. Just means it fails faster, no need to do the other stuff if this check fails.

timharrison-moj (approved):
LGTM

PR #205, opened 2025-01-07
PRC-386: Make reference code group an enum

james-baker-moj:
Also return 400 for an invalid group code instead of an empty list.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/resource/ReferenceCodesResourceIntegrationTest.kt
```diff
@@ -45,13 +48,23 @@ class ReferenceCodesResourceIntegrationTest : H2IntegrationTestBase() {
 
   @Test
   fun `should return empty list if no matching code found`() {
```
prabash-moj, line 50: this method name should be updated to 'fun `should return an error if no matching code is found`?

prabash-moj (approved):

PR #206, opened 2025-01-07
PRC-384 : Remove unnecessary fields from sync requests/responses

prabash-moj:
PRC-384 : Remove unnecessary fields from sync requests/responses


timharrison-moj (approved):
LGTM

PR #207, opened 2025-01-07
PRC-387: Standardise validation of reference codes

james-baker-moj:
Added missing validation for relationship type on creating contact.

Fixed issues with updating types when the reference code is no longer active (should be allowed).

Fixed issue with deleting phone number if the reference code for phone type was no longer active.


timharrison-moj (approved):
LGTM

PR #216, opened 2025-01-14
PRC-405: Add missing active flag on organisation

james-baker-moj:
Also includes the deactivated date which maps to expiry_date in NOMIS. In NOMIS, this is set to current date when you untick active and set to null if you reactivate.


timharrison-moj (dismissed):
LGTM

timharrison-moj (approved):
LGTM

PR #217, opened 2025-01-14
PRC-403: Only return global phone numbers in the phoneNumbers list for a contact.

timharrison-moj:
We were formerly including all address-specific numbers as well.
Address specific numbers are now only displayed and managed via editing the address they are associated with.


james-baker-moj (approved):
LGTM

PR #219, opened 2025-01-15
PRC-408: Initial migration endpoint for organisations

james-baker-moj:
Allow creation of an organisation using the NOMIS corporate id as the PK.

Allows for creation of all associated entities with it.

Currently does not handle duplicate migrations.

Remove use of @CreationTimestamp on organisation entities as this did not allow for overriding of the timestamp with one from NOMIS.


timharrison-moj (approved):
LGTM

PR #220, opened 2025-01-15
PRC-406: Handle duplicate organisations on migrate

james-baker-moj:
When a migration is run the client might time out or fail while the organisation is actually created. To allow for replays we must remove the existing organisation and any associated entities and then re-create them.

Also minor swagger doc fix.


timharrison-moj (approved):
LGTM

PR #222, opened 2025-01-16
PRC-407: Small modifications to contact migration

james-baker-moj:
Employments were actually already migrated. Changed to use employment table instead and move all the already migrated data to that table.

Also fixed issue with creation timestamps on migrated and synced contacts to ensure it's kept rather than overwritten.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/request/migrate/MigrateContactRequest.kt
```diff
@@ -203,8 +203,8 @@ data class MigrateEmployment(
   @Schema(description = "Unique sequence ID in NOMIS for this employment", example = "123")
   val sequence: Long,
 
-  @Schema(description = "The corporate organisation this person works for", nullable = true)
-  val corporate: Corporate? = null,
+  @Schema(description = "The corporate organisation this person works for")
+  val corporate: Corporate,
```
timharrison-moj, line 208: Does every employment have to have a corporate (org) linked? It makes sense, but just checking its always the case?

timharrison-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/request/migrate/MigrateContactRequest.kt
james-baker-moj: yeah there were no nulls

PR #223, opened 2025-01-16
PRC-401 : Tech ticket - Initial GET organisation API

prabash-moj:
PRC-401 : Tech ticket - Initial GET organisation API


src/main/resources/migrations/test/V2025.01.16.6__example_data.sql
```diff
@@ -0,0 +1,8 @@
+-- ==========================================================
+-- Example data
+-- Not loaded into any real environments - DEV, PREPROD or PROD
+-- Intended for integration tests and local-running only.
+-- ===========================================================
```
prabash-moj, line 5: Would the test environment not clear the DB before each run? so we could combine all the test SQLs into one file instead of maintaining versioned migration scripts for tests,

prabash-moj: I can be removed after we have create organisation endpoint

james-baker-moj: Yeah agree better to create using the API once we have it. I think `migrations/test` is only run in the test profiles so we can remove it safely then. I don't think the db is wiped between tests, we fire up the docker container only once and apply the migrations.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/resource/OrganisationController.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.facade.OrganisationFacade
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.model.response.City
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.model.response.Organisation
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.swagger.AuthApiResponses
+
+@Tag(name = "Organisation Data")
```
james-baker-moj, line 20: One tiny thing here. Could you make the tag "Organisation" and then add it to `OpenAPIConfiguration` after Restrictions please? It just helps keep the swagger organised. Really good job.

james-baker-moj (changes requested):

james-baker-moj (approved):
LGTM thanks

PR #224, opened 2025-01-17
PRC-241: Fix issues related to creating official contacts

james-baker-moj:
Fix validation of relationship to prisoner depending on the relationship type.

Fix missed rename of relationshipDescription

Return official contacts alongside social contacts from prisoner contact summary (this is for the prisoner contact list)


timharrison-moj (approved):
LGTM

PR #225, opened 2025-01-17
PRC-233: Return relationship type on PrisonerContactRelationshipDetails

james-baker-moj:
The type was missing and the relationship reference was looked up incorrectly when adding an official contact.


timharrison-moj (approved):
LGTM

PR #226, opened 2025-01-17
PRC-401 : Create organisation

prabash-moj:
PRC-401 : Create organisation


timharrison-moj (approved):
LGTM
Potentially, could be two different send() functions in outboundEventsService? One dealing with events related to Persons, and one related to Organisations? Avoids the nullable contact checks. We can wait and see how this develops though.

PR #230, opened 2025-01-21
PRC-419: Use org id and type as pk

james-baker-moj:
Change organisation_type to use organisation_id and organisation_type as the primary key instead of a surrogate key. This ensures no duplicates and removes the need to track ids for sync

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/repository/OrganisationTypeRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/repository/OrganisationTypeRepository.kt
@@ -5,10 +5,13 @@ import org.springframework.data.jpa.repository.Modifying
 import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
 import uk.gov.justice.digital.hmpps.hmppscontactsapi.entity.OrganisationTypeEntity
+import uk.gov.justice.digital.hmpps.hmppscontactsapi.entity.OrganisationTypeId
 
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/repository/OrganisationTypeRepository.kt
```diff
@@ -5,10 +5,13 @@ import org.springframework.data.jpa.repository.Modifying
...
 
 @Repository
-interface OrganisationTypeRepository : JpaRepository<OrganisationTypeEntity, Long> {
+interface OrganisationTypeRepository : JpaRepository<OrganisationTypeEntity, OrganisationTypeId> {
   @Modifying
-  @Query("delete from OrganisationTypeEntity c where c.organisationId = :organisationId")
+  @Query("delete from OrganisationTypeEntity c where c.id.organisationId = :organisationId")
   fun deleteAllByOrganisationId(organisationId: Long): Int
+
+  fun getByIdOrganisationId(organisationId: Long): List<OrganisationTypeEntity>
```
timharrison-moj, line 16: Should this be `getByOrganisationId`

timharrison-moj: Or is this getting by Id, but only one part of the primary key?

timharrison-moj (commented):
LGTM.

timharrison-moj (approved):
LGTM

PR #234, opened 2025-01-22
PRC-356: Deprecating the LANGUAGE, CITY, COUNTY, COUNTRY reference endpoints and service calls.

timharrison-moj:
This PR
* Replaces the use of language reference service calls in the contacts service with the common reference codes service.
* Deprecates the unused reference endpoints and services for CITY, COUNTY, COUNTRY and LANGUAGE.
* Adds the reasons for not using these data sets into the README file.


james-baker-moj (approved):
LGTM

PR #235, opened 2025-01-23
PRC-271: Organisation summary and search

james-baker-moj:
Basic search function for organisations which returns a list of summary objects.

Also fixed issue with contact address views using the wrong reference data and some additional padding that was in county reference data.

### DIFF
--- a/src/main/resources/migrations/common/R__v_contact_addresses.sql
+++ b/src/main/resources/migrations/common/R__v_contact_addresses.sql
@@ -17,12 +17,12 @@ select
     ca.street,
     ca.area,
     ca.city_code,
-    city_ref.nomis_description as city_description,
+    city_ref.description as city_description,
     ca.county_code,
-    county_ref.nomis_description as county_description,
+    county_ref.description as county_description,
     ca.post_code,
     ca.country_code,
-    country_ref.nomis_description as country_description,
+    country_ref.description as country_description,
     ca.verified,
     ca.verified_by,
     ca.verified_time,
@@ -36,9 +36,9 @@ select
     ca.updated_by,
     ca.updated_time
   from contact_address ca
-  left join city_reference city_ref on city_ref.nomis_code = ca.city_code
-  left join county_reference county_ref on county_ref.nomis_code = ca.county_code
-  left join country_reference country_ref on country_ref.nomis_code = ca.country_code
+  left join reference_codes city_ref ON city_ref.group_code = 'CITY' and city_ref.code = ca.city_code
+  left join reference_codes county_ref ON county_ref.group_code = 'COUNTY' and county_ref.code = ca.county_code
+  left join reference_codes country_ref ON country_ref.group_code = 'COUNTRY' and country_ref.code = ca.country_code
   left join reference_codes rc ON rc.group_code = 'ADDRESS_TYPE' and rc.code = ca.address_type;
 
 -- End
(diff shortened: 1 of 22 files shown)


src/main/resources/migrations/common/R__v_contact_addresses.sql
```diff
@@ -36,9 +36,9 @@ select
     ca.updated_by,
     ca.updated_time
   from contact_address ca
-  left join city_reference city_ref on city_ref.nomis_code = ca.city_code
-  left join county_reference county_ref on county_ref.nomis_code = ca.county_code
-  left join country_reference country_ref on country_ref.nomis_code = ca.country_code
+  left join reference_codes city_ref ON city_ref.group_code = 'CITY' and city_ref.code = ca.city_code
```
timharrison-moj, line 39: Eek. Good spot!

timharrison-moj: I searched for all models, entities, service usage etc... not direct table references (as we were not going to reference them directly). Anyway, gone now, all good.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/helper/TestAPIClient.kt
```diff
@@ -581,6 +596,20 @@ class TestAPIClient(private val webTestClient: WebTestClient, private val jwtAut
...
+    val totalElements: Int,
+    val first: Boolean,
+    val size: Int,
+    val number: Int,
+    val sort: ReturnedSort,
+    val numberOfElements: Int,
+    val empty: Boolean,
+  )
+
   data class ReturnedPageable(
```
timharrison-moj, line 613: Should this be in the main - `model/response` package rather than in a TestAPIclient?

timharrison-moj (approved):
LGTM. One question added.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/integration/helper/TestAPIClient.kt
james-baker-moj: It's just a test helper for things that return `Page`, which for some reason there is no common type defined in spring for us to use!  I guess we might need it if we ever call a 3rd party API with paging but for now it's only for tests

timharrison-moj: Thought the PageImpl does that? https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/PageImpl.html

timharrison-moj: PageImpl<T> - where T is the content row type. But fine for now! We will probably end up changing this area anway.

james-baker-moj: I couldn't get it to work as the `T` generic is lost when you define the `expectBody(PageImpl::class.java)` on `WebTestClient`. When I googled it people said "just create your own type..." so I did that. It's a similar problem to the swagger docs where we use `OrganisationSummaryResultItemPage` to generate the correct types. I also tried using `OrganisationSummaryResultItemPage` in `WebTestClient` but spring can't handle it. I suspect it doesn't provide the full constructor or sth.

PR #237, opened 2025-01-23
PRC-266: Add employment details to get contact by id

james-baker-moj:
Not the most efficient way of doing it as we load each organisation one at a time but given there are only a few employments per contact I think it's safe for now and I didn't want to over-optimise if we end up moving organisation data.


timharrison-moj (approved):
LGTM

PR #239, opened 2025-01-27
PRC-427: API to patch contact employments

james-baker-moj:
Support adding, updating and deleting employments all in a single API for the "edit employers" screen which allows staging multiple changes before submitting them.

Also discovered a small bug where fields that were Long or Boolean and required would actually default to 0 or false instead of returning an error.

Also moved the loading of organisations into `OrganisationService` so when we replace it a real service rather than in our DB it should be transparent to most of the code.


prabash-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/facade/EmploymentFacade.kt
```diff
@@ -0,0 +1,24 @@
...
+  private val employmentService: EmploymentService,
+  private val outboundEventsService: OutboundEventsService,
+) {
+
+  fun patchEmployments(contactId: Long, request: PatchEmploymentsRequest): List<EmploymentDetails> {
+    return employmentService.patchEmployments(contactId, request).also { result ->
+      result.createdIds.onEach { outboundEventsService.send(OutboundEvent.EMPLOYMENT_CREATED, it, contactId = contactId, source = Source.DPS) }
+      result.updatedIds.onEach { outboundEventsService.send(OutboundEvent.EMPLOYMENT_UPDATED, it, contactId = contactId, source = Source.DPS) }
+      result.deletedIds.onEach { outboundEventsService.send(OutboundEvent.EMPLOYMENT_DELETED, it, contactId = contactId, source = Source.DPS) }
+    }.employmentsAfterUpdate
```
timharrison-moj, line 22: This is nice and concise.

